PR #1009, opened 2021-04-06
use more recent version on github instead of sourcehut

### LINKED ISSUE #1008: Sourcehut repo from TROUBLESHOOTING is not updated
- OS Version: mac x86-64
- Browser Version: Google Inc. Chrome 86.0.4240.111
- Browser Addon Version: 0.2.4
- Neovim Plugin Version: 
- What I tried to do: troubleshoot with TROUBLESHOOTING.md
- What happened: the instructions say to clone the sourcehut repo into /tmp but it doesn't work with the current browser extension since that repo is not the most recent version of firenvim

It worked when I cloned the github repo instead.

jamespeapen:
The sourcehut repo has an older version that will not work with the
current browser extension version. 
Fixes #1008

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -64,7 +64,7 @@ Then, navigate to a page with a textarea (I really like `http://txti.es` for thi
 
 ## Make sure it's not your init.vim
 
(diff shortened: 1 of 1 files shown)


glacambre (approved):
Thank you for taking the time to improve the troubleshooting guide!

PR #1295, opened 2022-01-28
Consider E245 while parsing font name

tejasvi:
`:h E245` says
```vimhelp
- A '_' can be used in the place of a space, so you don't need to use
  backslashes to escape the spaces.
- Examples: >
    :set guifont=courier_new:h12:w5:b:cRUSSIAN
    :set guifont=Andale_Mono:h7.5:w4.5
```
As is indicated by [public dotfiles](https://cs.github.com/?q=%2F%20guifont.*_%2F%20language%3A%22Vim%20Script%22&scopeName=All%20repos&scope=), underscores are often used to represent spaces.

Also see https://github.com/glacambre/firenvim/issues/972#issuecomment-873635158

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -286,6 +286,7 @@ export function parseSingleGuifont(guifont: string, defaults: any) {
     if (defaults[fontFamily]) {
         result[fontFamily] += `, ${defaults[fontFamily]}`;
     }
+    result[fontFamily] = result[fontFamily].replaceAll('_', ' '); // :h E245
     return options.slice(1).reduce((acc, option) => {
             switch (option[0]) {
                 case "h":


glacambre (commented):
Thanks a lot for taking the time to look at firenvim's source code, improving it and sharing your change. I am very grateful you did all this.

However, I am not sure this change should be merged. The reason for that is that firenvim has to specify fonts using a name that will be recognized by the browser's CSS engine. [Underscores are allowed in CSS font names](https://www.w3.org/TR/css-fonts-4/#family-name-syntax) and because of that, there might be situations where the only way to refer to a font is to use an underscore, and replacing underscores with spaces would prevent users from referencing that font.

I agree that the current situation is not ideal (pretty terrible even, partly because of `guifont` being terrible and partly because of browsers rightfully attempting to protect user privacy). I wish I knew how to improve it :disappointed: .

tejasvi:
@glacambre No worries! Code was fun to read.

Issue #1642, opened 2024-10-06
Add Support for Inline Japanese Input in Firenvim


mtanda:
Hello!

First of all, I’d like to express my gratitude for creating such an incredibly useful tool. I recently started using Firenvim, and it has been extremely helpful for my workflow. I also appreciate that Japanese input is supported, which has made things much easier.

However, there’s a small improvement I’d like to suggest. While typing in Japanese, the text being input doesn’t appear "inline in the actual textarea" where it is meant to be entered. In my environment, the Japanese input system shows the text near the mouse cursor, but since it’s displayed slightly away from where I am typing, I find myself prone to making typing mistakes.

Therefore, I’d like to request support for "inline input." Similar requests have been raised in other tools, such as WezTerm, and they've successfully implemented it, as shown here: https://github.com/wez/wezterm/issues/3411

I tried making changes to support inline input myself, and it seems to work. If possible, could you please review my implementation?

Thank you once again for your hard work!

PR #1643, opened 2024-10-06
Add Support for Inline Japanese Input in Firenvim

mtanda:
will fix https://github.com/glacambre/firenvim/issues/1642

I utilized the existing `KeyHandler`. I made the background of the `KeyHandler` transparent and displayed it in front of the canvas to show the Japanese text being input. Normally, the display moves along with the mouse cursor, but I modified it to appear at the cursor position only while typing.

Since I haven’t fully explored the entire Firenvim implementation, there is a possibility that my changes might unintentionally break some existing functionality.

Before changes:
https://github.com/user-attachments/assets/fe63ff90-61ca-461f-af8d-33576dcefb2e

After changes:
https://github.com/user-attachments/assets/0c6a7277-e6f3-4183-90ba-97772b7df27e

### DIFF
--- a/src/KeyHandler.ts
+++ b/src/KeyHandler.ts
@@ -92,9 +92,11 @@ export class KeyHandler extends EventEmitter<"input", (s: string) => void> {
         if (isChrome()) {
             this.elem.addEventListener("compositionstart", (e: CompositionEvent) => {
                 this.isComposing = true;
+                this.elem.style.zIndex = "1";
             });
             this.elem.addEventListener("compositionend", (e: CompositionEvent) => {
                 this.isComposing = false;
+                this.elem.style.zIndex = "-1";
                 acceptInput(e);
             });
         }
--- a/src/index.html
+++ b/src/index.html
@@ -25,7 +25,7 @@
                 position: absolute;
                 margin: 0px;
                 padding: 0px;
-                z-index: 1;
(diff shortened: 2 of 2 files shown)


glacambre:
Thanks a lot for taking the time to implement this (and for recording videos - very helpful!). The only issue I'm seeing is that now the textarea's blinking caret is visible when moving the mouse around. As far as I can tell, adding `caret-color: transparent;` to the keyhandler's CSS is enough to fix it. Could you let me know if this has any adverse effects on your side?

mtanda:
Thank you for your feedback and for the suggestion! I’ve applied the fix by adding caret-color: transparent; to the keyhandler's CSS, and it appears that the blinking caret is now gone as expected.

To confirm, I’ve attached a new video that shows the updated behavior without the blinking caret issue. Please let me know if everything looks good or if there’s anything else I should address

https://github.com/user-attachments/assets/3f0a1c15-68bc-4628-a198-1487ee573c47

.

mtanda:
I’ve made an additional fix so that the textarea no longer intercepts mouse events while typing in Japanese. Previously, mouse events (such as right-click) were being intercepted, which prevented nvim’s context menu from displaying properly. Now, during Japanese input, mouse events are not intercepted, allowing nvim to handle them as expected.

PR #1653, opened 2024-12-15
Fix the NativeMessagingHosts path for Microsoft Edge

### LINKED ISSUE #1651: The NativeMessagingHosts/firenvim.json file is not installed for Microsoft Edge
- OS Version: macOS 15.1.1
- Browser Version: Microsoft Edge 131.0
- Browser Addon Version: 0.2.16
- Neovim Plugin Version: master (I suppose, commit 64f9389)

### What I tried to do

Followed the installation guide, installed both the extension and nvim plugin with lazy.vim:

```
  {
    "glacambre/firenvim",
    build = ":call firenvim#install(0)",
  },
```

Then opened a github issue, clicked the edit box, pressed cmd-e.

### What happened

nvim did not take over the interface.

Followed the troubleshooting guide and found out the reason is that the `firenvim.json` file is not created under

```
~/Library/Application\ Support/Microsoft\ Edge/NativeMessagingHosts/
```

I also have Chrome installed, so that file is present in
```
~/Library/Application\ Support/Google/Chrome/NativeMessagingHosts/
```

I copied it over and everything then works.

I took a glance of the `firenvim#internal_install` function, but the logic does not seem to be what I thought: it does iterate all known browsers and install the file for all of them. I have no ability to debug further so stopped looking. Maybe experts will tell easily.

Thanks for the awesome tool!

ymattw:
Fix #1651 for real :). Verified locally by command `nvim +'call firenvim#install(0)' +q`. Saw the firenvim.json file installed under `~/Library/Application Support/Microsoft Edge/NativeMessagingHosts`.

### DIFF
--- a/autoload/firenvim.vim
+++ b/autoload/firenvim.vim
@@ -499,7 +499,7 @@ function! s:get_edge_manifest_dir_path() abort
                 if !isdirectory(s:build_path(l:p))
                         let l:p = [$HOME, 'Library', 'Application Support', 'Microsoft Edge']
                 endif
-                return s:build_path([l:p, 'NativeMessagingHosts'])
+                return s:build_path(l:p + ['NativeMessagingHosts'])
         elseif has('win32') || s:is_wsl
                 return s:get_data_dir_path()
         end


glacambre:
D'oh! Thanks for the fix :)

PR #1668, opened 2025-01-09
firenvim.vim: supress error when run Vim (when config folder is the same)

vitaly-zdanevich:
My `~/.config/nvim` is a softlink to `~/.config/vim`, I do not want to see firenvim error when using Vim in a terminal.

### DIFF
--- a/plugin/firenvim.vim
+++ b/plugin/firenvim.vim
@@ -1,4 +1,8 @@
-if exists('g:firenvim_loaded')
-        finish
+if has('nvim')
+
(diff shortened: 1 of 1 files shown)


vitaly-zdanevich:
Oops looks like already solved in https://github.com/glacambre/firenvim/commit/9e83fcb62ea44d8f7b7935f17f864eae30abfddd ?

vitaly-zdanevich:
Yes, solved, thanks.

PR #1692, opened 2025-09-13
Testing farzadmf's mv3 MR

glacambre:


glacambre:
Firenvim is now ported to manifest v3, so I'll close this MR. Thanks again for the work you gifted me @farzadmf , and sorry I didn't actually use it.

PR #1710, opened 2026-04-29
drop preloaded instance

glacambre:
- background.ts: replace inmemory tab storage with browser.storage.session
- Survive mv3 background suspension destroying preloaded nvim instance
- Stop preloading instances entirely


glacambre:
[firenvim-0.2.16.zip](https://github.com/user-attachments/files/27198833/firenvim-0.2.16.zip)

PR #1712, opened 2026-05-01
manifestly v3

glacambre:
- background.ts: store all state in session storage
- Generate manifest v3 for chrome
- Chrome testsuite: disable coverage & tests requiring arbitrary js eval

### DIFF
--- a/src/testing/content.ts
+++ b/src/testing/content.ts
@@ -12,6 +12,6 @@ listenersSetup.then(() => {
     socket.addEventListener('message',
                             makeRequestHandler(socket,
                                                "content",
-                                               (new Function("return this"))().__coverage__
+                                               (globalThis as any).__coverage__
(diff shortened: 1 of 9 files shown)


src/testing/content.ts
```diff
@@ -12,6 +12,6 @@ listenersSetup.then(() => {
     socket.addEventListener('message',
                             makeRequestHandler(socket,
                                                "content",
-                                               (new Function("return this"))().__coverage__
+                                               (globalThis as any).__coverage__
```
justinmk, line 15: I did not know about the `new Function("return this")` trick, js is crazy 😂

glacambre: Yeah, it's pretty dirty... I should have used globalThis from the start, but I didn't know about it 5 years ago 😆

PR #1718, opened 2026-05-13
feat: support helium

sharpchen:
Add support for [helium browser](https://github.com/imputnet/helium)(chrome-based)

I've only tested on Windows.

[config path reference](https://github.com/imputnet/helium/issues/1670#issuecomment-4438958044)


glacambre:
LGTM - thanks!

PR #1724, opened 2026-07-21
Harden Lua server teardown and improve startup diagnostics

blacklight:
## Description

- `lua/firenvim.lua`: guard close_server against `nil` and already-closing server handles, switch from `vim.loop.close(server)` to `server:close()`, and pass the server argument in the `chunk == nil` handler. This prevents `vim.loop.close(nil)` from aborting shutdown and ensures `qall!` runs.
- `src/frame.ts`: log and propagate `writeFilePromise` rejections so `writefile/:edit` failures are visible in the frame console instead of being swallowed.
- `TROUBLESHOOTING.md`: document the LibreWolf/Firefox Local Network Access block on `ws://127.0.0.1` and the `network.lna.websocket.enabled=false` workaround (with the security/privacy trade-off noted).

## Test plan

- Build the extension with `npm run build`.
- Test Firenvim on a page with a `textarea`, on a Firefox version configured with `network.lna.websocket.enabled=false`; confirm the frame no longer crashes during server teardown.
- If possible, trigger a `writefile/:edit` failure and verify the error is logged in the frame console.
- On LibreWolf/Firefox with `network.lna.websocket.enabled=true`, confirm the documented LNA behavior and workaround.

### DIFF
--- a/lua/firenvim.lua
+++ b/lua/firenvim.lua
@@ -1,7 +1,9 @@
 local websocket = require("firenvim-websocket")
 
 local function close_server(server)
-        vim.loop.close(server)
+        if server ~= nil and not server:is_closing() then
+                server:close()
+        end
         -- Work around https://github.com/glacambre/firenvim/issues/49 Note:
         -- important to do this before nvim_command("qall") because it breaks
         vim.loop.new_timer():start(1000, 100, (function() os.exit() end))
@@ -32,7 +34,7 @@ local function connection_handler(server, sock, token)
         return function(err, chunk)
                 assert(not err, err)
                 if not chunk then
-                        return close_server()
+                        return close_server(server)
                 end
                 local _
                 if not headers then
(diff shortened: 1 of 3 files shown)


lua/firenvim.lua
```diff
@@ -32,7 +34,7 @@ local function connection_handler(server, sock, token)
         return function(err, chunk)
                 assert(not err, err)
                 if not chunk then
-                        return close_server()
+                        return close_server(server)
```
glacambre, line 37: Woops

lua/firenvim.lua
```diff
@@ -1,7 +1,9 @@
 local websocket = require("firenvim-websocket")
 
 local function close_server(server)
-        vim.loop.close(server)
+        if server ~= nil and not server:is_closing() then
```
glacambre, line 4: What are the remaining cases where server can be nil?

glacambre (commented):
Thanks for taking the time to make Firenvim better!

lua/firenvim.lua
blacklight: I considered cases where e.g. the server process may have not been started at all (e.g. prevented from the browser settings or crashed). But I couldn't explicitly replicate them, so consider this as an excess of precaution.

SuggonM:
Solves #1732

PR #1726, opened 2026-07-30
test: add the data dir on vimrc setup command for tests

Propfend:
I see duplicated line:
```js
fs.mkdirSync(nvim_conf_dir, { recursive: true });
```
where the data dir is not being created. This is a fix.


tests/_vimrc.ts
```diff
@@ -6,19 +6,30 @@ const process = require("process");
...
         process.env.XDG_CONFIG_HOME = path.join(base_dir, "config");
-        process.env.XDG_DATA_HOME = path.join(base_dir, "data");
+        process.env.XDG_DATA_HOME = nvim_data_dir;
+
         const nvim_conf_dir = path.join(process.env.XDG_CONFIG_HOME, "nvim");
+
         try {
                 fs.mkdirSync(nvim_conf_dir, { recursive: true });
-                fs.mkdirSync(nvim_conf_dir, { recursive: true });
+                fs.mkdirSync(nvim_data_dir, { recursive: true });
```
glacambre, line 23: Oof, thanks for catching that!

Do you think we should append `nvim` to `nvim_data_dir`, thhe same way we append `nvim` to `nvim_conf_dir`?

Propfend: yes, makes sense, why not.

PR #1728, opened 2026-07-30
test: add the headless mode support for chrome tests

Propfend:
Selenium already supports [running Chrome in headless mode](https://www.selenium.dev/blog/2023/headless-is-going-away/). This PR adds this option and uses the new way on the existing Firefox instantiation.


tests/firefox.ts
```diff
@@ -132,7 +132,7 @@ describe("Firefox", () => {
                 }
 
                 if (env["HEADLESS"]) {
-                        options.addArguments("-headless");
+                        options.addArguments("--headless=new");
```
glacambre: I can't find any documentation for this `=new` parameter, what does it do/mean?

Propfend: This is the new Chrome API:

Check the "What are the two headless modes?" section from the attached link the PR body:

image

PR #1734, opened 2026-09-01
fix(uv): prefer vim.uv with fallback

khaneliman:
## Summary

Prefer `vim.uv` for Firenvim's Lua server operations while retaining the
existing fallback.

## Why

[Neovim deprecates direct `vim.loop` use](https://neovim.io/doc/user/deprecated/#deprecated-0.10), while Firenvim supports older Neovim
releases.

## What changed

- Added one `vim.uv or vim.loop` binding.
- Migrated the five timer, pipe, and TCP calls to that binding.
- Preserved the existing server-handle cleanup behavior.

## Testing

- Luacheck passed with no warnings or errors.
- `npm run build` and `npm run lint` passed.
- Modern and fallback server startup probes passed.

## Related Issue(s)

- None.


glacambre:
Thanks a lot!

